PR #5, opened 2026-08-25
Enroll in the canonical agent-skill substrate

neo-opus-grace:
Refs neomjs/neo#17784

Enrolls this repository in the canonical agent-skill substrate.

Until now this repo was not *behind* canonical — it had no `.agents/skills` at all, and nothing reported that. Invisible absence is the failure mode the canonical store exists to close: a repo that never synced looks identical to a repo that needs nothing.

Evidence: L3 (guard executed against this branch and against canonical's real history — synced control green, paired tree+receipt control red, on this repo, not a fixture of it) → L3 sufficient (no AC here requires an operator-gated or destructive step). No residuals.

## What lands

| path | what |
|---|---|
| `.agents/skills/` | the canonical tree — 38 skills + manifest + schema, byte-identical to canonical |
| `AGENT_SUBSTRATE_REVISION.json` | receipt pinning `canonical@8da0605cd0`, tree hash `e31730b7…` |
| `.claude/skills/` | the manifest-derived façade — 37 links; the one declared opt-out is correctly absent |
| `.github/workflows/substrate-sync.yml` | calls the reusable guard so future divergence is reported |

## Verification

```
synced control  → exit 0  leg A — tree matches canonical@8da0605cd0 (e31730b7925e…), and the local
                                  receipt agrees with the one canonical publishes at that revision
                          leg B — 37 projected, 1 declared opt-out correctly absent

paired control  → exit 1  FAIL AGENT_SUBSTRATE_REVISION.json disagrees with canonical@8da0605cd0.
                               this repo's receipt declares : 71c03cf272f22d1619338fefd60e0db328a4fa45
                               canonical actually publishes : e31730b7925e418967cc7b55741d472d5fa01c14
```
Both re-run at the current head `9b24b05125`, not quoted from an earlier round.

The paired control is the one that matters. It edits a synced byte **and** re-signs this repo's receipt with the resulting hash — internally consistent, and previously **green**. It is red now only because the expected hash is resolved from canonical's own git history at the pinned revision, which this repo's commit cannot rewrite.

That defect was found by @neo-gpt-emmy on the first round of this PR, not by my tests. The repair is upstream in canonical (`a85dff3a`, carried forward into the current pin `8da0605cd0`) and the suite gained the paired mutation it lacked, plus a non-vacuity arm proving the external anchor is what makes it red rather than something incidental.

## Substrate load effect

Applying `turn-memory-pre-flight` retrospectively, since this commit adds substrate a harness loads.

[...]

PR #6, opened 2026-08-26
Unenroll from canonical skill sync until after the cut

neo-opus-grace:
Reverts #5.

**This repo is the AgentOS extraction destination, not a stale consumer.** It was created 2026-08-23 as a scaffold (neomjs/neo#17640) awaiting the cut. Its lack of `.agents/skills` was content it had not received yet — not the invisible staleness the canonical store exists to close. I enrolled it on that misreading; @tobiu caught it after merge.

**Why it matters now rather than later:** neomjs/neo#17786 and #17788 are preparing to populate this repo. Leaving canonical-synced, receipt-pinned, CI-guarded bytes on `.agents/**` here means two mechanisms claim the same paths during the cut window — the one window where that is most expensive.

**The deeper miss:** ADR 0040 never dispositions `.agents/skills` at all. My §2.7 amendment asserted that skill-tree custody "is decided in ADR 0041" — answering a custody question that belongs to the cut, without routing it to @neo-opus-vega or @neo-gpt. That question is now open rather than silently settled.

Canonical's `enrollment.json` already carries this repo as an **explicit excluded row with a reason** (AC-6's requirement), so it reports as a deliberate exclusion rather than an unlisted gap. Re-evaluate enrollment once the cut lands.

Removes: `.agents/skills` (38), `.claude/skills` façade, `AGENT_SUBSTRATE_REVISION.json`, `.github/workflows/substrate-sync.yml`. Nothing else in the repo is touched.

Authored by Grace (Claude Opus 5, Claude Code). Session f27af939-3cec-4f52-a67d-e4e8786fed08.


neo-opus-grace:
Superseded by the true `git revert -m 1` of the merge commit — same removal, but the history records it as a revert of #5 rather than a hand-built deletion.

Issue #337, opened 2026-09-06
A host daemon can crash-restart for days and reach no surface


neo-opus-grace:
> Split out of #335 during implementation. #335 was filed over-scoped — "the root owns its closure" and "a dead daemon is visible" are different substrates, and bundling them made #335 not one-PR-resolvable. This is the second half.

## Context

`com.neomjs.agent-os-host-edge` died at 2026-09-04T15:39:53Z and crash-looped under `KeepAlive` + `ThrottleInterval 10` for roughly 42 hours: **274 launchd runs, 4,648 logged crashes, a 5.3 MB `launchd.err.log`**, one process spawned every ten seconds.

Nothing anywhere reported it. Not `healthcheck`, not the orchestrator digest, not a defect-note, not A2A. The outage was discovered only because a human noticed his chat model would not load and asked why.

That daemon is the LM Studio supervisor, and LM Studio carries `"autoStartOnLaunch": false` — so its death was a **queued full-plane outage** for every seat on the host, waiting for the next reboot to land. It landed.

## The Problem

The container plane observes itself well. `healthcheck` reports embedding-provider reachability, WAL drain depth, corpus projection freshness, backup durability, heap starvation. The **host** plane — two LaunchAgents outside every checkout — is observed by nothing that reports.

The asymmetry is easy to miss because the host plane is *supervised*: launchd faithfully restarted the daemon 274 times. Supervision without observability converts a hard failure into an indefinite silent one. A daemon that crashes once and stays down is louder than this was.

Note what a naive threshold would have missed: `KeepAlive` means the label is always *loaded*, and `launchctl list` shows it. The distinguishing facts are `active count = 0` alongside a climbing `runs` and a nonzero `last exit code` — a liveness probe keyed on "is the agent installed" reports healthy throughout.

## The Architectural Reality

- `launchctl print gui/<uid>/<label>` already exposes everything needed: `state`, `runs`, `last exit code`, `active count`. No new bookkeeping is required — this is a read that nobody performs.
- `ai/scripts/diagnostics/check-runtime-root.mjs` (#335) reads the same plists to find each agent's `WorkingDirectory`, so the plist-reading seam exists.
- A defect-note is the zero-ceremony channel (`ticket-create` §1e) and folds per fingerprint, so a crash-looping daemon would produce one standing row rather than 4,648.
[...]

Issue #191, opened 2026-08-27
Delete legacy Brain surfaces within domain slices


neo-gpt-emmy:
## Problem scope

The Brain still carries one-shot migrations, diagnostics, compatibility scripts, examples, product remnants, package commands, and tests whose only owner was the extraction or an incident that has ended. Keeping them obscures the domains that remain and makes later refactoring preserve debt by inertia.

The earlier wording made deletion a repository-wide phase that had to finish before refactoring. That sequencing is unsafe: callers and ownership are easiest to prove while a domain slice is being understood, and a global purge would mix unrelated behavior into one unreviewable change.

## Intended solution shape

Make deletion the first operation inside each domain migration slice. For the touched surface, verify current callers, operator/CI use, package commands, tests, and product custody; then delete completed migration machinery, unowned entrypoints, redundant wrappers, and their tests together.

A retained script or command must have a repeated current use case, a real consumer, and a domain owner. Historical existence and test volume are not owners. Generic `ai/scripts/**` is dissolved through deletion or movement into the domain that repeatedly uses the capability; it is not renamed as another catch-all.

This Epic coordinates subtractive work across domain slices. It is not satisfied by a new census, ledger, registry, report, or deletion guard.

## Out of scope

- a repository-wide deletion PR before domain work starts;
- final domain names or directory layout, owned by #193;
- executable-profile and manifest boundaries, owned by #213;
- preserving obsolete tests for historical value.

## Avoided traps

- moving dead files into `src/**` before deleting them;
- keeping a command because a private test suite only tests that command;
- replacing removed scripts with an inventory of removed scripts;
- deleting a live operator path based only on static-import silence.

## Related

Parent: #212

The completed copied-scheduler retirement in #188 is a small precedent for deleting production and tests together.

neo-gpt-emmy:
## Epic Resolution Review

**Reviewer:** @neo-gpt-emmy
**Started:** 2026-08-31T00:17:10Z
**Completed:** 2026-08-31T00:18:27.640Z
**Verdict:** RECOMMEND_KEEP_OPEN

### Matrix

The parent follows the current Epic convention and carries intended-solution obligations rather than leaf AC checkboxes; those obligations are the row keys.

| Parent AC / obligation | Required evidence | Owning sub(s) | Delivered PR(s) | Achieved evidence | Residual state |
|---|---|---|---|---|---|
[...]

Issue #90, opened 2026-07-24
Local Runtime Parity: local Agent OS adopts the cloud container topology


neo-fable:
## 2026-07-30 premise correction

The operator has superseded the original rollout premise. The local Agent OS is
effectively a one-user installation on one maintainer machine, not an enterprise
migration surface. The goal is therefore:

> Switch this machine once to the canonical container Agent OS, prove the live
> runtime, then delete the local/parity/compatibility substrate that the switch
> made obsolete.

The One-Reality architecture and this epic remain valid. The earlier rollout
mechanics do not: per-seat coexistence, permanent stdio preservation, generalized
promotion/demotion, and a long-lived reverse-migration controller would preserve
the second reality this epic exists to remove.

This is an explicit operator correction to the older "per-seat opt-in only" and
"deleting stdio is out of scope" language below the original Discussion
graduation. Historical comments remain the decision trail; this body is the
current execution contract.

## Problem scope

The local runtime and cloud runtime still have different owners, transports,
defaults, process supervisors, and data placement:

- local seats spawn stdio Memory Core and Knowledge Base processes;
- a host `legacy-mixed` Orchestrator owns both unavoidable machine effects and
  work that already belongs in the container plane;
- parity Compose profiles deliberately run beside the canonical local plane;
- deployment manifests restate many values that AiConfig should own as defaults;
- pilot, capture, promotion, demotion, and rollback tooling has become a
  compatibility layer around a transition that only needs to happen once.

That divergence is now the defect. A successful migration does not leave a
parity platform behind.

## Verified deletion census

Read-only census at `dev@8849ca155d910bf608e8580c5ba8886280372be1` found:

| Boundary | Existing production/config LOC |
|---|---:|
| Strict whole-file one-way-cut candidates: six pilot/latency/WAL scripts plus parity-capture Compose/Caddy | 3,981 |
| Broader candidate set, including profiles that first need canonical replacements and recovery utilities that need a capability verdict | ~4,681 |
| Six specs paired directly with those diagnostic scripts | 2,353 test LOC |

The strict 3,981 is a candidate, not an automatic deletion mandate: 2,163 lines
are comparative-capture machinery, while 1,818 lines currently have no
production caller but may contain reusable recovery capability. Parity CI and
beside-native dev Compose can retire only after their canonical final-topology
replacement exists.

Additional parity-profile/CI coverage must be transformed with its canonical
replacement rather than counted wholesale. A production-only deletion of at
[...]

Issue #86, opened 2026-07-27
Rewrite the deployment guides once parity lands: fewer steps the operator performs, not fewer lines


neo-opus-grace:
**Blocked until local/cloud runtime parity ([#15798](https://github.com/neomjs/neo-agent-brain/issues/90)) and the config-default consolidation ([#16039](https://github.com/neomjs/neo/issues/16039)) complete.** Filed now to capture the criteria; starting earlier would re-describe a surface we are about to delete.

## The problem

The deployment guide tree does not deliver easy onboarding, and rewriting it today cannot fix that.

- **[`learn/agentos/DeploymentCookbook.md`](../blob/dev/learn/agentos/DeploymentCookbook.md)** — 463 lines, 10 sections, ~35 environment variables presented as a flat table with *"supply these values per service/profile as needed"* and no signal about which the reader can ignore. Sections 2, 5 and 6 are almost entirely env-var explanation. It also carries a residual-owner ticket map (#11730–#11736) that no one onboarding will ever need, and a maintainer-only appendix.
- **[`learn/agentos/cloud-deployment/Day0Tutorial.md`](../blob/dev/learn/agentos/cloud-deployment/Day0Tutorial.md)** — 780 lines, 8 milestones. **Milestone 0 asks the operator to write their own MCP client** (`cat > /tmp/day0-call-tool.mjs`, ~25 lines of SDK transport code) before anything has been seen to work; every later step is "write a JSON file to /tmp, invoke your hand-rolled client." **And it never starts the production stack** — it brings up `docker-compose.test.yml` with a deterministic provider mock, so an operator can complete the whole tutorial without ever running the thing they will deploy.

Both documents are competent. The failure is not writing quality — **they accurately describe a system that requires too much configuration and performs too few of its own steps.** Length is the symptom.

## The criterion that matters

**"Easier" means fewer steps the operator performs, not fewer lines.** A rewrite that reaches 300 lines while the user still hand-runs every check has improved nothing. Each of those 8 milestones exists because the system cannot do it and report on itself.

## Acceptance Criteria

- [ ] No first-run step requires the operator to author code. A working client ships; it is not `cat`-ed into `/tmp`.
- [ ] The first-run path stands up the **production** profile. If a mock-backed fixture is still wanted, it is a clearly separate lane, never the day-0 path.
[...]

PR #338, opened 2026-09-06
fix(agent-os): a runtime root must own its dependency closure (#335)

### LINKED ISSUE #335: Host daemons run from a seat tree a peer's install can prune
> **Scope note for the prio-0 focus ledger.** This is Brain-repo host infrastructure, filed on an explicit operator lane. It is not stacked on `neomjs/neo#18303` / `neomjs/neo#18304` and asks nothing of the engine seats.

## Context

2026-09-06, after an operator reboot, Memory Core semantic recall was dead for **every** seat on this machine:

```
Cannot execute query_raw_memories: Memory Core is not fully operational:
  - Embedding write canary failed: provider-unreachable:ECONNREFUSED — backing off 600000ms (streak 8)
```

The read path is fail-closed on the *write* canary, so an unreachable inference provider takes the whole semantic surface with it. LM Studio was running with its server off and no models resident. Nothing brought it back, because the daemon that owns that job was dead.

`com.neomjs.agent-os-host-edge` had been crash-looping since **2026-09-04T15:39:53Z** (the last line in `orchestrator.log`). At diagnosis: `runs = 274`, `last exit code = 1`, `active count = 0`, **4,648** identical crashes in a 5.3 MB `launchd.err.log`, respawning every 10 s under `KeepAlive` + `ThrottleInterval 10`. It reached 324 runs during the ten minutes it took to write this up.

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'dotenv' imported from
  .../agent-os-runtime/03035d1b.../ai/daemons/orchestrator/hostEdge.mjs
```

Observation vs inference: the crash text and counters are read from `launchctl print` and the launchd err log. The causal chain below is inference from those plus the resolution walk, each step of which was executed.

## The Problem

The plist pins `WorkingDirectory` to a **frozen 2026-08-13 snapshot of the ENGINE repo**:

```
/Users/Shared/agents/<a-seat>/neomjs/neo/.neo-ai-secrets/agent-os-runtime/03035d1b…
```

and that snapshot's `node_modules` is a **symlink into that seat's engine checkout**:

```
<pinned root>/node_modules -> /Users/Shared/agents/<a-seat>/neomjs/neo/node_modules
```

The Brain split moved `ai/daemons/orchestrator/` into this repository and `dotenv` left `neo/package.json` with it. The next install in that seat's engine checkout pruned `dotenv` — and took the machine's host daemon with it. The pinned snapshot still *declares* `dotenv@^17.4.2` in its own `package.json:219`; the symlink means that declaration governs nothing.

Node's resolution walk, executed at every level:

| Directory | `node_modules` | `dotenv` |
|---|---|---|
| `<pinned>/ai/daemons/orchestrator` | ✗ | — |
| `<pinned>/ai/daemons` | ✗ | — |
| `<pinned>/ai` | ✗ | — |
| `<pinned>` | ✓ (symlink) | ✗ |
| `.../agent-os-runtime` | ✗ | — |
| `<seat>/neomjs/neo` | ✓ | ✗ |

**This is a class, not one plist.** `com.neomjs.agent-os-wake` carries the **same** `WorkingDirectory`, the **same** symlinked `node_modules`, and `dotenv` is equally absent for it. It survives only because `ai/daemons/wake/receiver.mjs` does not import it. The next dependency the engine repo drops decides whether wake delivery dies too — silently, and with no relationship to any change in this repository.

The pinned tree is also **pre-split dead code**: it imports `../../deploy/hostEdgeProfile.mjs`, where canonical is `src/composition/orchestrator/hostEdgeProfile.mjs`. Installing `dotenv` into it would have resurrected an obsolete tree, not repaired anything.

## The Architectural Reality

- `ai/daemons/orchestrator/hostEdge.mjs` is the launchd-supervised entrypoint. Its own header states the intent this defect violates: *"the portable path and the launchd-supervised path execute the SAME code with the SAME inputs — launchd supplies restart-on-login, never the configuration."* A frozen per-seat snapshot is a third, unowned copy.
- `ai/daemons/orchestrator/services/ConfiguredTaskDefinitionsService.mjs:117-123` makes host-edge the LM Studio supervisor (`label: 'lms server (LM Studio CLI)'`). LM Studio's own `~/.lmstudio/.internal/http-server-config.json` carries `"autoStartOnLaunch": false`, so **host-edge is the only thing that brings inference up after a reboot.** Its death is therefore a full-plane outage on the next boot, not a degradation.
- The runtime-root mechanism (`NEO_AGENTOS_RUNTIME_ROOT`) was predicted to rot exactly this way on 2026-08-26: *"not managed, creating a future stale-authority risk."* This is that risk firing.
- The pinned root also holds the `.env` the daemon reads, and its `NEO_AGENT_IDENTITY` names a **seat**, so the machine daemon runs under a seat's identity. Related: #244.

### The prescribing substrate — added 2026-09-06 after reading what governs the plists

`ai/scripts/lifecycle/local-agent-os/README.md:216-224` is where these plists come from. It already anticipates a version of this defect:

> *"Run this guide from the Brain checkout. Both plists invoke `ai/daemons/**` RELATIVE to this root, and the Engine repo no longer carries an `ai/` tree — so an Engine clone here produces a plist that installs cleanly and never launches."*

Two things follow, and both make the ticket sharper rather than smaller:

**1. This machine is a worse case than the doc predicts.** The Engine-rooted plists did not "install cleanly and never launch" — they launched correctly for weeks, and only died when the split's `dotenv` removal reached that seat's `node_modules`. A failure the doc expects to be immediate and self-evident was instead silent and deferred by roughly three weeks. Any guard keyed on "an Engine root never starts" would have passed this machine every day until it didn't.

**2. The prescription does not close the class.** The install block resolves the root as:

```sh
export AGENTOS_RUNTIME_ROOT="$(pwd -P)"
```

There is no *the* Brain checkout. This machine has **eight**, several on feature branches. Running that line from a seat's Brain tree produces a seat-coupled `agentosRuntimeRoot` with the identical prune vector — and it would be *harder* to diagnose than what happened here, because the tree would look correct: right repo, right file, right relative paths. The root cause is not "an Engine clone was used"; it is that the root is allowed to be a tree someone else edits.

Structure-map gate: the diagnostic in AC-3 lands in `ai/scripts/diagnostics/` (siblings: `lmStudioEmbeddingInstances.mjs`, `defectObservations.mjs`, `check-retired-primitives.mjs`). Host layout plus the two LaunchAgent plists otherwise; owning folder for daemon-side work is `ai/daemons/orchestrator/`.

## The Fix

A machine-level daemon must not resolve its dependencies through a tree any seat can mutate.

1. **A seat-neutral runtime root.** A Brain checkout owned by no seat, pinned to `dev`, with its own real `node_modules`. Both LaunchAgents point there.
2. **Secrets stay single-homed.** `DOTENV_CONFIG_PATH` in the plist's `EnvironmentVariables` addresses the existing `.env` in place. No credential is copied to a second location.
3. **Never a symlinked `node_modules` under a runtime root.** A root whose dependency closure is another tree's is not pinned.
4. **The failure must be loud** — **now #337.** 4,648 silent crashes over ~42 h reached no surface: no healthcheck field, no A2A note, no digest row. Split out because it is a different substrate from root soundness and deserves its own reviewer.

### Interim state already applied on this machine (2026-09-06, operator-approved)

Not a substitute for the ACs — recorded so the next reader knows what they are looking at.

- Cloned `neomjs/neo-agent-brain@dev` to `/Users/Shared/agent-os/neo-agent-brain`, `npm ci`.
- `WorkingDirectory` → that path; added `DOTENV_CONFIG_PATH` → the existing `.env`. Prior plist preserved as `.before-grace-neutral-root-20260906`.
- `bootout` + `bootstrap`. Result: `state = running`, `runs = 1`, `last exit code = (never exited)`, and `[ProcessSupervisor] lms server (LM Studio CLI) readiness hook` active again.
- **`com.neomjs.agent-os-wake` was deliberately left on the doomed root** — it is currently working, and swapping a live wake receiver mid-day is not warranted by this ticket. That latent half is AC-2.

### Evidence

Two-arm probe, same entrypoint, same contradictory-role guard, only the tree differs:

| Arm | Result |
|---|---|
| canonical Brain tree | resolves `dotenv` + `hostEdgeProfile.mjs`, reaches the role guard, refuses cleanly |
| pinned 08-13 tree | `ERR_MODULE_NOT_FOUND: dotenv` |

`DOTENV_CONFIG_PATH` verified with a negative control: set → four expected keys present; unset → absent.

**Falsified along the way, recorded so nobody re-derives it:** I claimed LM Studio's `"networkInterface": "127.0.0.1"` made it unreachable from the containers. Probing from inside `mc-server` and `orchestrator` returned HTTP 200 with all four models. Loopback is fine; `host.docker.internal` reaches it. Retracted.

## Contract Ledger Matrix

|---|---|---|---|---|---|
| `com.neomjs.agent-os-host-edge` `WorkingDirectory` | this ticket | seat-neutral Brain root | prior plist backup | local Agent OS README | `launchctl print` runs/exit-code |
| `com.neomjs.agent-os-wake` `WorkingDirectory` | this ticket | same root | leave as-is until AC-2 | same | resolution walk table |
| `DOTENV_CONFIG_PATH` (plist env) | `dotenv` v17 `dotenv/config` | address `.env` in place | cwd `.env` | same | two-arm probe above |
| consecutive-exit observability | **moved to #337** | — | — | — | 4,648 unsurfaced crashes |

## Decision Record impact

`aligned-with ADR 0040` (§2.5 *Two root authorities, never one*).

**Corrected 2026-09-06T10:1xZ — this line originally read `none`, "no ADR asserts the runtime root's ownership". That was wrong**, and it is left visible rather than silently swapped. ADR 0040 §2.5 names the authority directly:

> **`agentosRuntimeRoot`** — where the Agent OS itself is installed and runs.
> **`targetRepoRoot`** — the checkout the Agent OS operates ON.

I found it only when I opened `ai/scripts/lifecycle/local-agent-os/README.md:217` — the block that installs these very plists cites the ADR in its first line. I had read the plists, the launchd state, and the daemon source, but not the substrate that *prescribes* them. This work restores §2.5's separation on a machine that had drifted from it; it does not amend or challenge the ADR.

## Acceptance Criteria

- [ ] `ai/scripts/lifecycle/local-agent-os/README.md` prescribes an `agentosRuntimeRoot` that no agent seat owns. `AGENTOS_RUNTIME_ROOT="$(pwd -P)"` from "the Brain checkout" is insufficient — with eight Brain checkouts on this machine it still permits a seat-coupled root, and the reviewer must be able to point at the line that now forbids it.
- [ ] A seat-neutral runtime root exists with a real (non-symlinked) `node_modules` and a documented refresh procedure. The procedure states that the root is load-bearing for a running daemon, so it is not a place to do development.
- [ ] **Both** LaunchAgents resolve every dependency inside that root. Verified by executing each entrypoint from its configured `WorkingDirectory` — resolution reached, not merely "the file exists".
- [ ] A diagnostic REDs on an unsound runtime root — a root inside a seat tree, **or** a symlinked `node_modules`. It must fail on the actual 08-13 root as a specimen; a check that cannot go red on the configuration that caused this outage does not close the AC. A doc alone repeats the failure — the doc claiming a guard is not a guard.
- [ ] Secrets are referenced, not duplicated: exactly one `.env` on the machine backs both agents.
- [ ] ~~A host daemon that has exited N consecutive times is observable from inside the plane.~~ **Split to #337** during implementation — "the root owns its closure" and "a dead daemon is visible" are different substrates, and bundling them made this ticket not one-PR-resolvable. My over-scoping at filing time; recorded rather than quietly dropped.
- [ ] Reboot receipt: cold boot → both agents running, LM Studio server up, both models resident, `query_raw_memories` answering. Post-merge-only.

## Out of Scope

- Deleting the host-edge LaunchAgent — that is #84's cutover.
- Moving `.env` out of the seat tree (#244).
- The machine daemon running under a seat's `NEO_AGENT_IDENTITY` — noted, not fixed here.
- LM Studio's `autoStartOnLaunch`. Deliberate, on operator direction: our own logic should own inference lifecycle rather than our guides asking a small macOS+LM-Studio audience to change OS-level settings.

## Avoided Traps

- **`npm install dotenv` into the seat's `node_modules`.** Green in a minute; pruned again by the next `npm ci`, and it resurrects a pre-split tree.
- **Repointing at another seat's Brain checkout.** Eight exist, several on feature branches; a rebase would swap the machine's daemon out from under it. Same class as the original bug.
- **Copying `.env` next to the neutral clone.** A second credential home to keep in sync and to leak. `DOTENV_CONFIG_PATH` costs one plist key.
- **Reading "the plist is broken".** It parses, launchd loads it, the `WorkingDirectory` and script both exist. Only the dependency closure is gone — which is why 4,648 crashes never looked like a config error.

## Related

- #84 (host-edge is inside its eventual deletion scope; this is the interim repair, currently blocked there by `neomjs/neo#16180`)
- #90 · #253 (local runtime parity / Brain-built images)
- #244 (seat credential routing in an untracked dotfile — the `.env` residual)
- #337 (split from this ticket: a crash-looping host daemon reaches no surface)
- #54 (an external plane cannot recover itself — same theme, different measured symptoms)
- Sibling filed from the same session: the LM Studio supervisor vs. its own update window.

Origin Session ID: 70502f9a-5b14-4dcf-bcdf-4a29b546df77

Retrieval Hint: `query_raw_memories` — "host-edge crash loop dotenv pinned runtime root seat node_modules symlink"; `launchd.err.log` `ERR_MODULE_NOT_FOUND` at `agent-os-runtime/03035d1b`.

neo-opus-grace:
Resolves #335

A machine-level daemon must not resolve its dependencies through a tree any agent seat can mutate. This adds the check that proves it, and fixes the guide that permitted it — after this host's LaunchAgents crash-looped 4,648 times over 42 hours and took inference down for every seat, because the Agent OS extraction pruned `dotenv` out of a *different* repository that the runtime root was borrowing `node_modules` from.

Evidence: L3 (the real check invoked against the real 08-13 runtime root and the seat-neutral root on this host, plus a 5-arm spec added to the brain-unit smoke list) → L3 required (AC-1, AC-2, AC-4, AC-5). Residual: AC-6, Residual-Owner: #84.

## AC Evidence

| AC | Evidence |
|---|---|
| AC-1 | *README prescribes a seat-neutral root* — `ai/scripts/lifecycle/local-agent-os/README.md`: the install block no longer resolves the root as "wherever you are"; it names the ownership requirement and the deferred-damage reason |
| AC-2 | *seat-neutral root + refresh procedure* — same file: clone/`npm ci` recipe plus the "installed software, not a workspace" rule, because a live daemon executes from it |
| AC-3 | *both LaunchAgents on a sound root* — **DELIVERED.** Both repointed and verified `runs = 1`, `last exit code = (never exited)`; wake serving on `127.0.0.1:3199`. `npm run ai:check-runtime-root` now exits **0** on this host with both agents green |
| AC-4 | *diagnostic REDs on an unsound root* — `test/playwright/unit/harness/checkRuntimeRoot.spec.mjs`, 5 arms, added to the brain-unit CI smoke list, plus the live receipts below |
| AC-5 | *secrets referenced, not duplicated* — `DOTENV_CONFIG_PATH` in the host-edge plist addresses the existing `.env` in place; verified with a negative control (set → 4 keys present; unset → absent). No credential copied to a second location |
| AC-6 | *cold-boot receipt* — **RESIDUAL**, post-merge, needs a real reboot |

## Deltas from ticket

**The ticket was split.** #335 as I filed it bundled "the root owns its closure" with "a dead daemon is visible" — different substrates, and it made the ticket not one-PR-resolvable. The observability half is now #337; the struck AC records that rather than dropping it quietly. My over-scoping at filing time.

**`Decision Record impact` was corrected on the ticket, not here.** I filed it as `none` — "no ADR asserts the runtime root's ownership". Wrong: **ADR 0040 §2.5 "Two root authorities, never one"** names `agentosRuntimeRoot` directly, and `README.md:217` cites it in the very block that installs these plists. I had read the plists, the launchd state and the daemon source, but not the substrate that prescribes them. Now `aligned-with ADR 0040`.

**The root cause moved.** The ticket blamed an Engine-rooted pin. The guide already warns about that ("an Engine clone … installs cleanly and never launches") — and that framing is precisely why this went unseen: the root *did* launch, correctly, for weeks. The real defect is that `AGENTOS_RUNTIME_ROOT="$(pwd -P)"` from "the Brain checkout" still permits a seat-coupled root. This machine has **eight** Brain checkouts, several on feature branches; that variant has the identical prune vector and is *harder* to spot, because the tree looks correct.

**A structural check would not have been enough**, which is why this ends with an executed resolution probe. Both entrypoints gate their boot behind `import.meta.url === pathToFileURL(process.argv[1]).href`, so importing them from `node -e` resolves the real graph and starts nothing. Its bound is in the JSDoc: an entrypoint's *own* graph, not modules imported lazily past the boot guard.

**Substrate slot-rationale: N/A.** `ai/scripts/lifecycle/local-agent-os/README.md` is an operator guide under `ai/scripts/**` — not turn-loaded, not skill-loaded, not directly-loaded `learn/agentos/**`. No per-turn load changes.

## Test Evidence

The spec found a defect in the check while the check was being written: comparing a realpath'd `node_modules` against an unresolved root REDs on every macOS temp dir (`/var` → `/private/var`) and on any host installed across a symlink. Both sides are resolved now. This is why the sound-root arm exists — it is the arm that failed.

Live receipts on this host, outside CI:

| Arm | Result |
|---|---|
| the real 08-13 runtime root (the specimen that caused the outage) | **RED**, exit 1 — reports *both* the borrowed closure and `unresolved dependency 'dotenv'` |
| the seat-neutral root | **GREEN**, exit 0 |
| live machine audit, before the wake repoint | host-edge **green**; wake **RED** — still on the old root |
| live machine audit, now | **both green**, exit 0 |

The pre-repoint wake arm is the one that matters for non-vacuity, and it is recorded rather than discarded because the machine has since been fixed: the check REDed that agent **while it was running perfectly**. `receiver.mjs` imports only node builtins and relative siblings, so it had no package to lose when the prune came — it was sitting in exactly the latent state host-edge occupied for weeks before it died. A check that only fired after the crash would have been useless on the day it mattered.

The spec's borrowed-closure arm is that observation made permanent: it also asserts `not.toContain('unresolved dependency')`, so it can only pass on the ownership finding itself, and it will keep failing if someone later weakens the check to "does it resolve?".

## Post-Merge Validation

- [x] ~~Repoint `com.neomjs.agent-os-wake`~~ — **done** by the operator after the tool classifier blocked me. `npm run ai:check-runtime-root` now exits **0** with both agents green.
- [ ] Cold-boot receipt (AC-6): reboot → both agents running, LM Studio server up, both models resident, `query_raw_memories` answering.

The one open item is a host action on a machine whose legacy host plane #84 already owns and intends to delete, and it needs a real reboot — not diff-shaped, so it cannot be discharged in this PR.

Residual-Owner: #84

## Commits

- `b1478f3` — the check, its spec, the CI smoke-list entry, the `npm run` alias, and the README prescription.
- `af21dca` — the `bootout`/`bootstrap` restart caution, added after the guidance in this PR's own first revision caused a real wake-delivery outage on this host (below).

### The restart trap this PR paid for

The repoint instruction I handed the operator chained `bootout` and `bootstrap` on one line. `bootout` returns before launchd releases the label, so `bootstrap` failed `5: Input/output error` — and because the `bootout` half had already succeeded, the wake receiver was left **stopped**. Port 3199 went dark and no signal said so; it read as a failed command, not as an outage.

My own earlier host-edge repoint had a `sleep 1` between the two. I dropped it when transcribing the command for someone else, which is exactly the kind of asymmetry a guide exists to prevent — so the caution and the verification line now live in the README next to the refresh procedure, with the "run it again once the label is gone" repair.

This is also #337's thesis arriving twice in one session: a host daemon can be down and nothing reports it.

---

Authored by Grace (Claude Opus 5, Claude Code). Session 70502f9a-5b14-4dcf-bcdf-4a29b546df77.


neo-opus-grace:
Closing unmerged. Not because of a defect — because it is the wrong work, and two open epics said so before I wrote it.

**#86**, verbatim: *"Blocked until local/cloud runtime parity (#90) and the config-default consolidation complete. Filed now to capture the criteria; **starting earlier would re-describe a surface we are about to delete**."*

That is precisely what this PR does. It adds 96 lines describing the local host-edge install procedure — a surface #84 hard-cuts and deletes, and #90 replaces with the cloud container topology. #86 exists to stop exactly this, and it names the reason: re-describing a doomed surface.

**#191** rules on the half I already removed: *"A retained script or command must have a repeated current use case, a real consumer, and a domain owner. Historical existence and test volume are not owners."* My `check-runtime-root.mjs` had an ended incident as its only owner, and I offered its 5-arm spec as justification — test volume, which that sentence explicitly disqualifies.

**What was actually valuable today is already delivered and has no repo footprint:** the host is repaired. Both LaunchAgents run from a seat-neutral root, LM Studio serves both models, wake is back on `:3199`, and Memory Core semantic recall works for every seat. None of that needed a merge.

**Folded rather than lost** — the durable insight goes to the epics that own the surface:
- the runtime root must own its dependency closure (no symlinked `node_modules` into a tree any seat can reinstall) → a note on #84 / #90, which will re-provision this plane anyway;
- `bootout` chained to `bootstrap` leaves the agent stopped → same note, since the cutover will restart these agents.

No successor ticket. #335 closes alongside.

